PR #110, opened 2021-02-09
Update to mamba=0.7.12

xhochy:
We currently ship an older mamba release which has issues with private channels, updating to the latest release fixes these.

I added an explicit pin for the mamba version to improve reproducibility.

### DIFF
--- a/Miniforge3/construct.yaml
+++ b/Miniforge3/construct.yaml
@@ -26,7 +26,7 @@ specs:
 {% endif %}
 
 {% if name.startswith("Mambaforge") %}
-  - mamba
+  - mamba 0.7.12
 {% endif %}
   - conda {{ version.split("-")[0] }}
 


hmaarrfk:
the explicit pin will cause us to forget to move it forward.

is >= better?

wolfv:
maybe not pin a version on ppc64le because sometimes the builds fail on Travis. I can be more careful to not merge when they do from now on ...

xhochy:
Actually I would like to have the pin to detect that we don't support a current mamba release on all platforms. Like the CI result tells us currently.

I would prefer to keep the pin and rather have a cron job that opens a PR that updates the pin instead. Not sure whether we should do that via Github Actions or the cf-scripts (cc @conda-forge/bot).

xhochy:
Using `mamba` to get a better error message doesn't help.
```
$ CONDA_SUBDIR=linux-ppc64le mamba create -n mambatest mamba=0.7.12 pypy

Looking for: ['mamba=0.7.12', 'pypy']

conda-forge/noarch       Using cache
conda-forge/linux-ppc64l [====================] (00m:00s) Done
Encountered problems while solving.
Problem: nothing provides python 3.6.9 1_73_pypy needed by pypy-7.3.1-h9f0ad1d_1
```

xhochy:
CI is green, we should either merge this or continue the discussion on the mamba pinning.

xhochy:
@conda-forge/miniforge Can we tag this as a release? That would be nice as I run into a bug with custom channels and the `setup-miniconda` action with `mamba` here. I have workarounds but a tag would clean my workflows up greatly.

PR #185, opened 2021-07-16
Add linting action with shellcheck and according problem matcher

dbast:
This just visualises the findings via problem matcher in the Github diff view and the shellcheck exit code is currently ignored.

The majority of the current code in this repo currently are shell scripts. Linting those with shellcheck helps to review new contributions.

If this gets accepted, I can fix the current findings as part of this PR or a succeeding one to get the noise down.


.github/shellcheck-matcher.json
```diff
@@ -0,0 +1,17 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "shellcheck",
+      "pattern": [
+        {
+          "regexp": "^(.+):(\\d+):(\\d+):\\s(note|warning|error):\\s(.*)\\s\\[(SC\\d+)\\]$",
+          "file": 1,
+          "line": 2,
+          "column": 3,
```
dbast, line 10: This would only match to warnings and errors and ignore *notes* ... that would be maybe better as less noise.
```suggestion
          "column": 3,
          "severity": 4,
```

hmaarrfk:
the double quoting might not be bad practice. i do often find myself annoyed at people adding spacing in directory names. i think that would address that problem

PR #199, opened 2021-08-13
use newest micromamba

wolfv:
Checklist

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -38,7 +38,7 @@ cp LICENSE $TEMP_DIR/
 ls -al $TEMP_DIR
 
 if [[ "$TARGET_PLATFORM" != win-* ]]; then
-    MICROMAMBA_VERSION=0.13.0
+    MICROMAMBA_VERSION=0.15.2
     mkdir $TEMP_DIR/micromamba
     pushd $TEMP_DIR/micromamba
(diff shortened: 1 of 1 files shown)


xhochy:
We should setup a Github Action to auto-update that, I can do that soon.

wolfv:
Thanks @xhochy that would hopefully allow us to catch these errors a bit earlier :( need to look into what's breaking here.

scripts/build.sh
```diff
@@ -39,7 +39,7 @@ cp LICENSE "${TEMP_DIR}/"
 ls -al "${TEMP_DIR}"
 
 if [[ "${TARGET_PLATFORM}" != win-* ]]; then
```
ericpre, line 41: Now that https://github.com/mamba-org/mamba/issues/961 is fixed, it should be possible to use micromamba on windows too?

This was disabled in https://github.com/conda-forge/miniforge/pull/160, because of https://github.com/conda-forge/miniforge/issues/158.

ericpre: I will try in a virtual machine.

ericpre: Great, sounds good!
Removing menu when uninstalling doesn't work, see https://github.com/conda/constructor/blob/f14fa91e1e363a029fffccb74c81618d27285bfb/constructor/nsis/main.nsi.tmpl#L981

I don't know if other arguments are missing but just in case, the `conda-standalone` arguments for `constructor` are defined in https://github.com/conda-forge/conda-standalone-feedstock/blob/374ae2c27efcf5fd5cd38ddea316f59ff242067f/recipe/entry_point.py#L30-L51

PR #232, opened 2021-11-19
Update miniforge version used in installer for OSX

hmaarrfk:
Checklist

### DIFF
--- a/build_miniforge_osx.sh
+++ b/build_miniforge_osx.sh
@@ -16,12 +16,13 @@ elif [ "$(uname -m)" = "arm64" ]; then
 else
    exit 1
 fi
-curl -s -L $condapkg > miniconda.sh
+curl -s -L "$condapkg" > miniconda.sh
 sha256sum miniconda.sh | grep $conda_chksum
 
 bash miniconda.sh -b -p ~/conda
 
 echo "Configuring conda."
+# shellcheck disable=SC1090
 source ~/conda/bin/activate root
 
 export CONSTRUCT_ROOT="${PWD}"
--- a/build_miniforge_osx.sh
+++ b/build_miniforge_osx.sh
@@ -10,8 +10,12 @@ miniforge_arch="$(uname -m)"
 miniforge_version="4.10.3-10"
 condapkg="https://github.com/conda-forge/miniforge/releases/download/${miniforge_version}/Mambaforge-${miniforge_version}-MacOSX-${miniforge_arch}.sh"
 if [ "$(uname -m)" = "x86_64" ]; then
+   # hmaarrfk: 2021/12/01 Not too sure how to check sha256sums on OSX
+   # shellcheck disable=SC2034
    conda_chksum="7c44259a0982cd3ef212649678af5f0dd4e0bb7306e8fffc93601dd1d739ec0b"
 elif [ "$(uname -m)" = "arm64" ]; then
+   # hmaarrfk: 2021/12/01 Not too sure how to check sha256sums on OSX
+   # shellcheck disable=SC2034
    conda_chksum="72bc86612ab9435915b616c2edb076737cbabe2c33fd684d58c2f9ae72e1957c"
 else
    exit 1
--- a/build_miniforge_osx.sh
+++ b/build_miniforge_osx.sh
@@ -22,7 +22,6 @@ sha256sum miniconda.sh | grep $conda_chksum
 bash miniconda.sh -b -p ~/conda
 
 echo "Configuring conda."
-# shellcheck disable=SC1090
 source ~/conda/bin/activate root
 
 export CONSTRUCT_ROOT="${PWD}"
--- a/build_miniforge_osx.sh
+++ b/build_miniforge_osx.sh
@@ -4,14 +4,15 @@ set -e
 set -x
 
 echo "Installing a fresh version of Miniforge3."
-MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/download/4.8.3-1"
-MINIFORGE_FILE="Miniforge3-4.8.3-1-MacOSX-x86_64.sh"
+MINIFORGE_VERSION="4.10.3-10"
+MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}"
+MINIFORGE_FILE="Mambaforge-${MINIFORGE_VERSION}-MacOSX-x86_64.sh"
 curl -L -O "${MINIFORGE_URL}/${MINIFORGE_FILE}"
 bash "${MINIFORGE_FILE}" -b
 
 echo "Configuring conda."
 # shellcheck disable=SC1090
-source ~/miniforge3/bin/activate root
+source ~/mambaforge/bin/activate root
 
 export CONSTRUCT_ROOT="${PWD}"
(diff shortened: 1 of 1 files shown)


build_miniforge_osx.sh
```diff
@@ -4,14 +4,15 @@ set -e
 set -x
 
 echo "Installing a fresh version of Miniforge3."
-MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/download/4.8.3-1"
-MINIFORGE_FILE="Miniforge3-4.8.3-1-MacOSX-x86_64.sh"
+MINIFORGE_VERSION="4.10.3-8"
+MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}"
+MINIFORGE_FILE="Mambaforge-${MINIFORGE_VERSION}-MacOSX-x86_64.sh"
```
jakirkham, line 9: Should the `Mambaforge` portion also be refactored out?

build_miniforge_osx.sh
```diff
@@ -4,14 +4,15 @@ set -e
...
 curl -L -O "${MINIFORGE_URL}/${MINIFORGE_FILE}"
 bash "${MINIFORGE_FILE}" -b
```
jakirkham, line 11: Should we set the path with `-p` here so that paths below don't need to be updated?

jakirkham (commented):
Thanks Mark! 😄 Had a couple questions below

build_miniforge_osx.sh
hmaarrfk: I'm not sure how the `-p` flag helps here. Can you elaborate? Will it avoid the command

```
source ~/mambaforge/bin/activate
```
or will it make it explicit that we are installing it at the directory `~/mambaforge`

build_miniforge_osx.sh
hmaarrfk: I'm not sure that really helps. The reason I refactored the version number was to make it easier to spot in 2 lines that needed to keep it consistent. Here we already have a variable called `MINIFORGE_FILE`. I'm not sure what else we can do.

build_miniforge_osx.sh
jakirkham: It could be used to call this something generic like `conda_install`. Admittedly somewhat cosmetic, but it also saves us changing another line if we change this again in the future

jakirkham, line 11:
```suggestion
CONDA_INSTALLER="Mambaforge"
CONDA_INSTALLER_VERSION="4.10.3-8"
CONDA_INSTALLER_URL="https://github.com/conda-forge/miniforge/releases/download/${CONDA_INSTALLER_VERSION}/${CONDA_INSTALLER}-${CONDA_INSTALLER_VERSION}-MacOSX-x86_64.sh"
curl -L "${CONDA_INSTALLER_URL}" > ~/conda_installer.sh
bash ~/conda_installer.sh -b -p ~/conda_install
```

build_miniforge_osx.sh
```diff
@@ -4,14 +4,15 @@ set -e
...
 
 echo "Configuring conda."
 # shellcheck disable=SC1090
-source ~/miniforge3/bin/activate root
+source ~/mambaforge/bin/activate root
```
jakirkham, line 15:
```suggestion
source ~/conda_install/bin/activate root
```

jakirkham (commented):
To provide more context to the earlier comments, was thinking something like this. So in the future we really only need to change a couple environment variables

hmaarrfk:
Honestly, I do kinda see these repos as places where I would want to point other users to on how to "use" mamba/miniforge.

I don't think I want to install things to a "non-standard" directory yet given that we don't need to. The extra level of indirection is really annoying for new readers.

I think it is a bit of a shame that we can't find a common name for mambaforge and miniforge.

hmaarrfk:
I would prefer avoiding a custom installation location, and avoiding an extra variable if we can.

jakirkham:
How is it non-standard? Users are free to set the install path wherever is convenient. In some cases (particularly on systems where the home directory uses NFS), the default is not even a good idea.

hmaarrfk:
I guess non-standard is the wrong word. It just seems "different" than what most people need.

The variable isn't really used beyond this 5 line script. So there is really no point in making it one in my mind.

Again, the main motivation in creating a variable for the version number is it bring it to the "front of the line of code" as opposed to the end of it highlighting its importance.

hmaarrfk:
I think you are right. I just want to keep in sync with the anvil images. So I'm trying to follow 
https://github.com/conda-forge/docker-images/blob/main/scripts/run_commands

hmaarrfk:
seems to pass osx. i hope i got the uname correct,

build_miniforge_osx.sh
```diff
@@ -4,14 +4,31 @@ set -e
...
+elif [ "$(uname -m)" = "arm64" ]; then
+   # hmaarrfk: 2021/12/01 Not too sure how to check sha256sums on OSX
+   # shellcheck disable=SC2034
+   conda_chksum="72bc86612ab9435915b616c2edb076737cbabe2c33fd684d58c2f9ae72e1957c"
+else
+   exit 1
+fi
+curl -s -L "$condapkg" > miniconda.sh
+# hmaarrfk: 2021/12/01 Not too sure how to check sha256sums on OSX
+# sha256sum miniconda.sh | grep $conda_chksum
```
jakirkham, line 25:
```suggestion
openssl sha256 miniconda.sh | grep $conda_chksum
```

jakirkham, line 25:
```suggestion
openssl sha256 miniconda.sh | grep $conda_chksum
```

build_miniforge_osx.sh
```diff
@@ -4,14 +4,31 @@ set -e
...
+# Keep variable names in sync with 
+# https://github.com/conda-forge/docker-images/blob/main/scripts/run_commands
+miniforge_arch="$(uname -m)"
+miniforge_version="4.10.3-10"
+condapkg="https://github.com/conda-forge/miniforge/releases/download/${miniforge_version}/Mambaforge-${miniforge_version}-MacOSX-${miniforge_arch}.sh"
+if [ "$(uname -m)" = "x86_64" ]; then
+   # hmaarrfk: 2021/12/01 Not too sure how to check sha256sums on OSX
+   # shellcheck disable=SC2034
+   conda_chksum="7c44259a0982cd3ef212649678af5f0dd4e0bb7306e8fffc93601dd1d739ec0b"
+elif [ "$(uname -m)" = "arm64" ]; then
```
jakirkham, line 16: @erykoff is this correct for macOS ARM? Or should we have something else here?

jakirkham:
Seems reasonable. Thanks for the updates Mark 😄

Suggested a change for `sha256` (this should work on other platforms as well as long as they have `openssl`)

Not sure about macOS ARM, but asked someone who may know more above

build_miniforge_osx.sh
erykoff: Running `uname -m` on my machine does indeed say `arm64`.

jakirkham: Thanks for checking! 😄

jakirkham (approved):

PR #237, opened 2021-11-26
Add instructions to use rosetta in a environment

Haydnspass:
Incorporate instructions to use rosetta
https://github.com/conda-forge/miniforge/issues/165#issuecomment-860233092

Checklist
not applicable because README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,15 @@ export DOCKERIMAGE=condaforge/linux-anvil-aarch64
 bash build_miniforge.sh
 ```
 
+## Rosetta on Mac with Apple Silicon Hardware
+To create new environments that use packages compiled for Intel chips (using rosetta) one should use `CONDA_SUBDIR=osx-64`.
+This environment variable should be used upon creation and permanently set in the respective environment:
+```bash
+CONDA_SUBDIR=osx-64 conda create -n [environment]   # create a new environment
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -194,6 +194,15 @@ export DOCKERIMAGE=condaforge/linux-anvil-aarch64
 bash build_miniforge.sh
 ```
 
+## Rosetta on Mac with Apple Silicon Hardware
+To create new environments that use packages compiled for Intel chips (using rosetta) one should use `CONDA_SUBDIR=osx-64`.
```
xhochy, line 198: Shouldn't this be in the conda docs instead of here?

hmaarrfk: Maybe it should go here: https://github.com/conda-forge/conda-forge.github.io/blob/main/src/user/tipsandtricks.rst

Haydnspass: @hmaarrfk 
I don't mind where I just think that there would be a benefit of documenting it :) The reason why I put it in here is because many M1 users will read some medium articles which typically point to miniforge.

So is tipsandtricks the decision or are there other opinions?

davidxia:
seems to be closed by https://github.com/conda-forge/conda-forge.github.io/pull/1576

hmaarrfk:
Yes. Thank you!

PR #306, opened 2022-05-17
README.md modified for avoiding prematured closing of the shell.

sktomar:
Checklist

Following the steps given in **Uninstall Process**, it is possible that a user might close the shell before the 3rd step is completed. In this case, the command `conda` won't be found in the path, and a full path needs to be given. A simple solution is to ignore closing the shell before all the 3 steps are completed.


hmaarrfk:
thanks.

PR #375, opened 2022-11-01
Readme: Note that latest installers use Python 3.10 and PyPy 3.9

EwoutH:
In #333 the default environments were updated to use Python 3.10 and PyPy 3.9. This commit updates that in the Readme.


hmaarrfk:
thank you!

PR #382, opened 2022-11-10
Use temporary default GITHUB_TOKEN for update PRs

dbast:
By extending the scope of the per run temporaray default GitHub action token `GITHUB_TOKEN` the need of using ssh keys for cloning/pushing updates is eliminated.

So no need to manage an extra ssh anymore and this is also a security improvement as ssh keys could be leaked, while the token is only valid during the workflow runs.

This got possible via the changes announced here a while ago: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

### DIFF
--- a/.github/workflows/autoupdate.yml
+++ b/.github/workflows/autoupdate.yml
@@ -2,13 +2,18 @@ name: Auto-update mamba
 on:
  schedule:
    - cron: "0 */6 * * *"
+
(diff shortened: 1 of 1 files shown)


xhochy (changes requested):
With this change CI would not be triggered anymore on PRs generated by the workflow.

dbast:
@xhochy Oh, I wasn't aware about the CI trigger issue. Thanks!

PR #478, opened 2023-08-21
add small FAQ

jaimergp:
Comes from discussion at https://github.com/conda-forge/conda-forge.github.io/pull/1995#issuecomment-1685967253

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,21 @@ export DOCKERIMAGE=condaforge/linux-anvil-aarch64
 bash build_miniforge.sh
 ```
 
+## FAQ
+ 
+### What's the difference between Mambaforge and Miniforge?
+    
+After the release of Miniforge 23.3.1 in August 2023, Miniforge and Mambaforge are essentially identical. The only difference is the name of the installer and subsequently the default installation path.
+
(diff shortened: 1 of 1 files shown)


ocefpaf (approved):
Perfect!

README.md
```diff
@@ -229,6 +229,21 @@ export DOCKERIMAGE=condaforge/linux-anvil-aarch64
 bash build_miniforge.sh
 ```
 
+## FAQ
+ 
+### What's the difference between Mambaforge and Miniforge?
+    
+After the release of Miniforge 23.3.1 in August 2023, Miniforge and Mambaforge are essentially identical. The only difference is the name of the installer and subsequently the default installation path.
+
+Before that release, Miniforge only shipped conda, while Mambaforge added mamba on top. Since Miniconda started shipping conda-libmamba-solver in July 2023, Miniforge followed suit and started shipping it too in August. At that point, since conda-libmamba-solver depends on libmambapy, the only difference between Miniforge and Mambaforge was the presence of the mamba Python package. To minimize surprises, we decided to add mamba to Miniforge too.
```
hmaarrfk, line 238: is the July 2023 thing true? I only just set the 23.3.0 and 23.3.1 as latest

jaimergp: Anaconda published those Miniconda installers in July 2023, containing conda 23.5. CalVer can make this conversation confusing I guess 😬, but I spelt out the month and year hoping people wouldn't understand it as a version, but simply as a date. If there's a different way we can write it out, I am super happy to change it.

jakirkham: Maybe we should bump the installer version for clarity?

https://github.com/conda-forge/miniforge/pull/480

PR #482, opened 2023-08-24
Try menuinst v2 (DNM)

jaimergp:
Just debugging what would happen if we were to release menuinst v2 stable in the wild.

Do not merge and sorry for the noise! 🙏


jaimergp:
Errors to be fixed by https://github.com/conda/menuinst/pull/151

jaimergp:
Windows errors are only on PyPy because the canaries are not building for that interpreter. The C extension is essentially the same in v1 and v2 so I don't anticipate any problems when get to the feedstock.

jaimergp:
Causes behind errors:

* Windows: conda >=23.3 has a menuinst constraint <2, so the solver fails to solve (as it should)
* osx-arm64, linux-aarch64, linux-ppc64le: menuinst 2.0 not yet available there. Requested migration already.

PR #489, opened 2023-09-07
DOC: README.md: updates for Windows users

mdhaber:
xref gh-453

Adds commentary about Windows installation and usage. Describes implications of adding path environment variable and mentions the Miniforge Prompt added to the Start menu.

Feel free to push changes directly. For example, perhaps use of `C:\Users\myusername\miniforge3\condabin` is not recommended. (However, it seems to work for me, and there is little risk of conflict between those files and other software besides the obvious, like Anaconda.)

Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,11 @@ rm -fr ${HOME}/.conda
 
 ### Windows
 
-Download the installer and double click it on the file browser.
+Download and execute the Windows installer. Follow the prompts, taking note of the options to
+"Create start menu shortcuts" and "Add Miniforge3 to my PATH environment variable". The latter is
+not selected by default due to potential conflicts with other software. Without Miniforge3 on the
+path, the most convenient way to use the installed software (such as commands `conda` and `mamba`)
+will be via the "Miniforge Prompt" installed to the start menu.
 
 ### Non-interactive install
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -133,7 +133,11 @@ rm -fr ${HOME}/.conda
 
 ### Windows
 
-Download the installer and double click it on the file browser.
+Download and execute the Windows installer. Follow the prompts, taking note of the options to
+"Create start menu shortcuts" and "Add Miniforge3 to my PATH environment variable". The latter is
+not selected by default due to potential conflicts with other software. Without Miniforge3 on the
+path, the most convenient way to use the installed software (such as commands `conda` and `mamba`)
+will be via the "Miniforge Prompt" installed to the start menu.
```
hmaarrfk, line 140:
```suggestion
will be via the "Miniforge Prompt" installed to the start menu.

There are known issues with the usage of special characters and spaces in
the installation location, see for example
https://github.com/conda-forge/miniforge/issues/484.
We recommend users install in a directory without any such characters in the name.
```

hmaarrfk:
I think this is a step in the right direction. I don' t want to needlessly delay this good addition. thanks!

PR #573, opened 2024-03-26
update macOS installation instructions

fakerybakery:


README.md
```diff
@@ -95,18 +95,29 @@ Latest installers with Mamba and PyPy in the base environment:
...
 or
 
-    wget "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
-    bash Miniforge3-$(uname)-$(uname -m).sh
+```bash
+wget "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
+bash Miniforge3-$(uname)-$(uname -m).sh
+```
+
+**NOTE:** For macOS, you may need to run `bash` with `sudo`:
```
hmaarrfk, line 115: I'm not sure this is correct.

We actually prefer that users not use sudo unless they really know what they are doing, then they don't need us to tell them to use sudo. The preferred way is simply to install in your user's home directory like the default will do.

Can you expand on the usecase where you find `sudo` necessary or desired?

hmaarrfk: it should be attempting to change your personal zshrc which you should have access to.

That is, it should be trying to edit (on OSX)
```
/Users/fakerybakery/.zshrc
```

which should be a file you control, that is writable by your user.

It maybe that you are running into a real bug.

If you can recreate the traceback and the error that would be really useful.

I don't have an OSX machine, but others might be able to help troubleshoot.

fakerybakery: I got:
```
TypeError: memoryview: a bytes-like object is required, not 'str'
```

jakirkham: This is likely a bug in the installer code. If you can share more of the traceback (or better yet how the install was done), that would help in identifying where this comes from

Issue #656, opened 2024-10-05
Maintenance of the PyPy installers


hmaarrfk:
### Comment:

Now that PyPy has been deprecated from conda-forge, is there still any energy behind maintaining the PyPy installers?

jakirkham:
Think this is covered under the same depreciation

We can drop them

PR #657, opened 2024-10-06
Deprecate PyPy3 installers

hmaarrfk:
Gonna leave this up here for 1 week for comments ahead of cutting a release:

Closes https://github.com/conda-forge/miniforge/issues/656

Checklist

### DIFF
--- a/Miniforge3/pypy_deprecation.sh
+++ b/Miniforge3/pypy_deprecation.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+if [[ "$GITHUB_ACTIONS" == "true" ]]; then
+    echo "::warning title=Mambaforge is now deprecated!::Future Miniforge releases will NOT build Mambaforge installers. We advise you switch to Miniforge at your earliest convenience. More details at https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/."
+else
+    echo "!!!!!! Conda-Forge PyPy is now deprecated !!!!!"
+    echo "Future Miniforge releases will NOT build with PyPy in the base environment installers."
+    echo "We advise you switch to Miniforge3 at your earliest convenience."
+    echo "More details at https://conda-forge.org/news/2024/08/14/sunsetting-pypy/"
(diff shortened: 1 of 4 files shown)


Miniforge3/pypy_deprecation.sh
```diff
@@ -0,0 +1,27 @@
...
+    echo "Future Miniforge releases will NOT build with PyPy in the base environment installers."
+    echo "We advise you switch to Miniforge3 at your earliest convenience."
+    echo "More details at https://conda-forge.org/news/2024/08/14/sunsetting-pypy/"
+    echo "If you are unable to switch to Miniforge, you may pin your installer version to one found in "
+    echo "https://github.com/conda-forge/miniforge/releases/tag/24.3.0-1"
+    echo "or if you lack the system requirements (Linux glibc >= 2.17, or macOS + x86-64bit >= 10.13)"
+    echo "you may pin your installer to one older version found in "
+    echo "https://github.com/conda-forge/miniforge/releases/tag/24.3.0-0"
+    echo "This Miniforge installer will ceese to work in 2025."
+fi
```
jakirkham, line 16: Let's put this in a text file that ships with the installer. Then all platforms can echo the test file and we can get the wording to match with both

jakirkham (commented):
Thanks Mark! 🙏

Added a suggestion below

Miniforge3/pypy_deprecation.sh
hmaarrfk: I would rather not overbuild for a depreation that will only last 3 months.

The challenge is that the wording  MUST be different for different platforms.

Windows, linux, and OSX have different recommendations.

Generic recommendations REALLY confuse first time users.

Is there a word here that you are concerned with getting correct?

hmaarrfk: My windows skills are also seriously lacking..... so I would rather not try something new....

My main goal for this week is to give time for somebody that might care for this installer to stepup, but not to create more organizational work. "2" notices is not enough to reorganize things IMO.

PR #686, opened 2024-11-27
Improve readme

### LINKED ISSUE #685: README.md quite confusing: reorganisation needed
I participate (as a teacher) to a Python training. There were many Windows and macOS users and the README page leads to confusions.

We gave this link https://github.com/conda-forge/miniforge. I now realize that we should have given https://conda-forge.org/download/ which may be better (but actually not really because it lacks very important information to get a correct setup).

Most users reading the text in https://github.com/conda-forge/miniforge want to install the latest version of miniforge.
Currently the explanation how to do that is spread in different parts of the README. There are sections Download, Installation and Usage.

A Windows user first reads the section Download and then do not read the rest of the README. He/she clicks on https://github.com/conda-forge/miniforge/releases or on [latest](https://github.com/conda-forge/miniforge/releases/latest) in the table and obtain a page which is not very useful with plenty of links with complicated names. The deprecated names starting with Mambaforge or Miniforge-pypy3 appear first. In practice most participants to the training were completely lost.

For Windows the interesting link is now https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Windows-x86_64.exe

It seems to me that the README should be reorganized with a first section Install with subsections:

- Windows
- Unix-like platforms (macOS & Linux)
- macOS
- Specificity for CI pipeline

Commands and actions needed to complete the installation should also be mentioned.  There should be in this section enough information to get a decent installation with at least `condabin` in the path (not only with Miniforge Prompt, which is of course very bad in 2024 since Windows users can and should use a better shell and a better terminal).

So we need some worlds on `conda init` and `conda config --set auto_activate_base false`.

I will try to submit a PR to improve the situation.

paugier:
Fix #685.

I try to reorganize the README so that users can more easily find the information they need to install Miniforge.

The result can be read here https://github.com/paugier/miniforge/tree/improve-readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,57 @@ You can still find the latest installers in the [24.9.2 release](https://github.
 
 ## Usage
 
-If Miniforge is on the system path (default on macOS and Linux), its versions of the
-[`conda`](https://conda.io/projects/conda/en/latest/user-guide/getting-started.html#managing-environments) and
-[`mamba`](https://mamba.readthedocs.io/en/latest/user_guide/mamba.html#mamba-user-guide) programs can be used
-at any command prompt. The most notable difference is that the default channel for packages will be conda-forge.
+Miniforge provides installers for the commands [`conda`](https://conda.io/) and
+[`mamba`](https://github.com/mamba-org/mamba).  Once the installer for your OS
+and architecture has been executed, you should be able to use these commands in
+a terminal.
 
+### conda/mamba usable in any terminals
+
+However, with the default choices of the Windows installer, these commands are
+only available in the "Anaconda Prompt". To be able to use these commands in
+other terminals, one needs to initialize conda for your shell by running in
+the Anaconda Prompt
+
+```sh
+conda init
+```
+
+Note that one can also just add the `C:\Users\myusername\miniforge3\condabin` folder
+to the path environment variable
+[manually](https://learn.microsoft.com/en-us/previous-versions/office/developer/sharepoint-2010/ee537574(v=office.14)#to-add-a-path-to-the-path-environment-variable)
+so `conda` and `mamba` may be used more conveniently from any command prompt with limited
+chance of software conflicts.
+
+The same situation arises on Unix if you use the non-interactive install.
+Initialization can be done by calling conda with its full path, with something like
+
+```sh
+~/miniforge3/bin/conda init
+```
+
+### Automatic activation of environments
+
+By default, once conda has been initialized for your shell, the `base` environment is
+activated so that the command `python` corresponds to the base Python provided by
+Miniforge and `conda install` installs packages in the `base` environment. This can
+be convenient for but it is cleaner to deactivate this automatic activation with
+
+```sh
+conda config --set auto_activate_base false
+```
+
+and use `conda` or `mamba` to create and activate other environments, with for example
+(to create an environment called `main`)
+
+```sh
+conda create --name main jupyterlab numpy pandas
+conda activate main
+```
+
+Finally, it is also possible to add to your shell configuration file
+(typically `~/.bashrc` or `~/.zshrc` on Unix;
+on Windows, edit with `notepad $PROFILE`) the activation command.
 
 ## Requirements and installers
 
@@ -119,20 +165,7 @@ The versions listed as "System: 32-bit" are not compatible with the installers o
 
 `(***)` Apple silicon builds are experimental and haven't had testing like the other platforms.
 
-
 ## Install
-
-On all platforms, the installation should be finalized with commands like
-
-```sh
-conda init --all
-```
-
-(see `conda init -h`) and potentially also
-
-```sh
-conda config --set auto_activate_base false
-```
 
 ### Windows
 
@@ -142,11 +175,6 @@ Follow the prompts, taking note of the options to
 not selected by default due to potential conflicts with other software. Without Miniforge3 on the
 path, the most convenient way to use the installed software (such as commands `conda` and `mamba`)
 will be via the "Miniforge Prompt" installed to the start menu.
-
-If desired, the
-`C:\Users\myusername\miniforge3\condabin` folder may be added to the path environment variable
-[manually](https://learn.microsoft.com/en-us/previous-versions/office/developer/sharepoint-2010/ee537574(v=office.14)#to-add-a-path-to-the-path-environment-variable)
-after installation so the software may be used more conveniently from any command prompt with limited chance of software conflicts.
 
 There are known issues with the usage of special characters and spaces in
 the installation location, see for example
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,138 +86,195 @@ You can still find the latest installers in the [24.9.2 release](https://github.
...
-Download and execute the Windows installer. Follow the prompts, taking note of the options to
+Download and execute [the Windows installer](https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Windows-x86_64.exe).
+Follow the prompts, taking note of the options to
 "Create start menu shortcuts" and "Add Miniforge3 to my PATH environment variable". The latter is
 not selected by default due to potential conflicts with other software. Without Miniforge3 on the
 path, the most convenient way to use the installed software (such as commands `conda` and `mamba`)
 will be via the "Miniforge Prompt" installed to the start menu.
 
+If desired, the
+`C:\Users\myusername\miniforge3\condabin` folder may be added to the path environment variable
```
paugier, line 147: Note that for installation for all users, the path is of course different. It would be good to mention it.

paugier:
Note that I don't think the failure for miniforge / Miniforge-pypy3-Linux-aarch64 has anything to do with this PR.

README.md
```diff
@@ -113,138 +86,195 @@ You can still find the latest installers in the [24.9.2 release](https://github.
...
 </details>
 
-## Install
 
-### Unix-like platforms (macOS & Linux)
+## Usage
 
-Download the installer using curl or wget or your favorite program and run the script.
-For eg:
+If Miniforge is on the system path (default on macOS and Linux), its versions of the
```
hmaarrfk, line 92: I want to change the whole idea of "Miniforge" being on the path.

"Miniforge" gets "conda" and "mamba" on your system. Once it is installed, its "job" is done. I want to help emphasize this.

hmaarrfk (commented):
I know I've seen you around, so I know you have the best intentions.

But as is, this diff moves too many pieces around and is difficult for me to review. 

Maybe somebody else can help review it, but I have limited bandwidth for this.

I want to in general move people away from this page since this isn't the right place to discuss "general conda problems".

paugier:
> I know I've seen you around, so I know you have the best intentions.

My goal is to improve this README which is a real issue since I saw that many users get lost and do not manage to use the installer correctly.

I don't care about this PR. It might be bad. However, I'm sure that this file has to be deeply reorganized so that people with a limited experience in conda-forge should be able in 2 minutes to install miniforge correctly and get conda/mamba in their terminals.

> But as is, this diff moves too many pieces around and is difficult for me to review.

This PR moves and reorganizes texts and Github is not very good are showing such changes.

One should really compare

- before : https://github.com/conda-forge/miniforge
- after : https://github.com/paugier/miniforge/tree/improve-readme

and think about the user experience of a user with no strong knowledge on conda, conda-forge, Github releases pages and co.

> I want to in general move people away from this page since this isn't the right place to discuss "general conda problems".

Then what would be the right page for people wanting to install and setup conda/mamba with miniforge?

hmaarrfk:
> One should really compare

This helps alot. thanks I didn't think of that.


> install and setup conda/mamba with miniforge?

"Install" is one thing. This is the right place for this. "The bash script Miniforge.sh crashed" should be addressed here.

"Setup" is a different thing that starts to emcompass the old `.condarc` file. I don't know where a good place for this is.

For example, this user https://github.com/conda-forge/miniforge/issues/690 has `anaconda's channel` (well a mirror of it) included so I'm not sure how to start helping.

I think we are using https://github.com/conda-forge/conda-forge.github.io as a "catchall repository". However in https://github.com/conda-forge/miniforge/issues/602 i made the point that this repo had more stars.

README.md
```diff
@@ -113,138 +86,195 @@ You can still find the latest installers in the [24.9.2 release](https://github.
...
-```bash
-CONDA_BASE_ENVIRONMENT=$(conda info --base)
-echo The next command will delete all files in ${CONDA_BASE_ENVIRONMENT}
-# Warning, the rm command below is irreversible!
-# check the output of the echo command above
-# To make sure you are deleting the correct directory
-rm -rf ${CONDA_BASE_ENVIRONMENT}
+## Install
+
+On all platforms, the installation should be finalized with commands like
```
hmaarrfk, line 125: i don't think this is true on unix. 

You should instead explain what the defaults will do, and explain why you suggestion is important.

README.md
hmaarrfk: thank you.

README.md
```diff
@@ -113,138 +86,195 @@ You can still find the latest installers in the [24.9.2 release](https://github.
...
 
-    start /wait "" Miniforge3-Windows-x86_64.exe /InstallationType=JustMe /RegisterPython=0 /S /D=%UserProfile%\Miniforge3
+Run the script with:
 
-### Downloading the installer as part of a CI pipeline
+```sh
+bash Miniforge3-$(uname)-$(uname -m).sh
+```
+
+or for non-interactive install (call with `-h` to list the extra options):
```
hmaarrfk, line 182:
```suggestion
The interactive installation will prompt you to add initialize conda with your shell. This is typically with recommended workflow. 

For non-interactive install (for example on a CI), the following command can be used (call with `-h` to list the extra options):
```

README.md
```diff
@@ -113,138 +86,195 @@ You can still find the latest installers in the [24.9.2 release](https://github.
...
+
+```sh
+bash Miniforge3-$(uname)-$(uname -m).sh -b
+```
+
```
hmaarrfk, line 187:
```suggestion

In non-interactive installations, the conda initialization commands will not be run by default.
```

README.md
```diff
@@ -113,138 +86,195 @@ You can still find the latest installers in the [24.9.2 release](https://github.
...
+### macOS with Homebrew
+
+On macOS, you can install miniforge with [Homebrew](https://brew.sh/) by running
+
+```sh
+brew install miniforge
+```
+
```
hmaarrfk, line 195: I would rather remove this section. I'm happy to point to it, but I would rather omit the installation instructions and add a disclaimer that users should turn their issues to homebrew prior to asking us to do so.

hmaarrfk (commented):
Thank you I think this is a step in the right direction. Just a few comments I would like your thoughts on.

paugier:
Thanks for the review @hmaarrfk! I'm not a good English writer but I tried to add clear explanations in e7201d3388dcd8d4d08ba7383afb2896d688d3ed.

Again, it is easier to see what it gives in https://github.com/paugier/miniforge/tree/improve-readme

paugier:
Note for myself: once we converge on something good here, it would be good to add few lines in https://conda-forge.org/download/

paugier:
Happy new year @hmaarrfk. Could these new changes be reviewed.

hmaarrfk:
ok thanks.

PR #711, opened 2025-01-04
Remove miniforge name customization

hmaarrfk:
As part of shedding more customization to keep this more maintainable

Merge after https://github.com/conda-forge/miniforge/pull/709


hmaarrfk:
While this is optional, i feel like it is important to remove these kinds of indirection to help reduce the maintenance burden and improve readability of our scripts for future users.

Issue #730, opened 2025-02-28
Maintaining relevance in 2025


hmaarrfk:
### Comment:

I'm wondering if there is more we can do to shed maintenance burden in 2025

- [ ] Decreasing the test matrix for the test images. We likely don't need to test for Ubuntu 16.04 for example: https://github.com/conda-forge/miniforge/blob/main/build_miniforge.sh#L19
    - Should keep us more nimble
- [ ] Removing the Darwin name from the OSX installers: https://github.com/conda-forge/miniforge/blob/main/.github/workflows/ci.yml#L97
    - Should shed about a few installers and mental load from the "downloads page"
- [ ] Dropping `mamba`
    - This is more controversial, but with `conda-libmamba-solver` the performance of conda has greatly improved.
    - With the recent challenges getting `mamba 2` rolled out I can't say I'm too excited to try again.
    - perhaps we have a transition packages where `mamba` is a pointer to `conda` but emits a warning when used? 

Followup of https://github.com/conda-forge/miniforge/issues/602

jakirkham:
Probably latest and LTS are enough for Ubuntu. Though would be interested to hear from others

Maybe same story with Debian (or is testing on Ubuntu enough?)

CentOS 7 testing makes sense given how much we use it in builds

Know the goal is shedding, but it may make sense to test UBI 8 (effectively a limited RHEL 8 in a container), AlmaLinux 8 and/or Rocky Linux 8 given the current EL landscape. Think we need to test something that is not EOL, but could see wide use

<hr>

One less installer name with the same functionality? Sounds like a good change

<hr>

Am hopefully that after the initial churn `mamba` 2 becomes more stable and friendly. The Mamba team fixed a lot of issues in that upgrade. Think that should be lauded. Wouldn't want teams to feel reluctant to fix complex issues that need breaking changes because of potential reception

Cadair:
> * Removing the Darwin name from the OSX installers

I'm 👎 on this because of the nice install instructions which tell people to run `bash Miniforge3-$(uname)...` which gives you Darwin on mac. See https://github.com/conda-forge/conda-forge.github.io/pull/2573

hmaarrfk:
> the nice install instructions which tell people to run bash Miniforge3-$(uname)... w

yeah.... this is quite a strong opinion.

PR #738, opened 2025-03-12
Rebuild miniforge installation scripts

jjerphan:
`fmt 11.1` broke the ABI, breaking some builds of packages, including the one of spdlog and mamba 2.0.7.

In the meantime new builds of mamba have been issued to fix the problem.

See: https://github.com/mamba-org/mamba/issues/3862

### DIFF
--- a/Miniforge3/construct.yaml
+++ b/Miniforge3/construct.yaml
@@ -1,4 +1,4 @@
-{% set version = os.environ.get("MINIFORGE_VERSION", "25.1.1-1") %}
+{% set version = os.environ.get("MINIFORGE_VERSION", "25.1.1-2") %}
 {% set conda_libmamba_solver_version = "25.1.1" %}
 # This file is parsed by the scripts to define
 #   - `MICROMAMBA_VERSION` in `scripts/build.sh`


SylvainCorlay:
This looks good to me.

SylvainCorlay:
> Do we need to pull: https://github.com/conda-forge/miniforge/releases/tag/24.11.3-1

I think so, unless we are fine shipping miniforge with packages marked as broken.

( I think we should have marked fmt 11.1.* as broken until an ABI-compatible patch release is available upstream).

hmaarrfk:
well could this PR then please revert mamba2 as well?

SylvainCorlay:
> well could this PR then please revert mamba2 as well?

Why?

hmaarrfk:
as i mentionned, i've personally had trouble with the mamba1 -> mamba2 transition on a few of my machines quite literally yesterday. It has just been a very tiring transition that has made me consider dropping `mamba` from Miniforge3 alltogether (I don't want to do it, but I think it is something to consider given the amount of recent churn it has caused me personally). https://github.com/conda-forge/miniforge/issues/730#issue-2886133892

I understand that this is a fresh install, but if the build from `24.11.3-1` is problematic, then I want to make sure we release 1 build in March 2025 (with all the fixes that @jjerphan pushed to constructor) for mamba1.

Please don't get me wrong, I'm extremely happy you and @jjerphan are helping more recently with Miniforge, and it is a main reason why I'm backing off from my proposal of dropping `mamba` from Miniforge.

However, I think it would be great to have 2 release:

1. One with mamba1
2. One with mamba2 

in quick succession to help revert in case we need to.

jjerphan:
https://github.com/conda-forge/miniforge/pull/737/files#diff-faa2e4718e45104bcbb7bbd2a41537c729a7e1d2994128d3527379d13b1978c4L1 recreated installers for 1.5.12 with the latest version of constructor.

This PR recreates installers for the second builds of 2.0.7 (which are sane because built for fmt<11.1) with the latest version of constructor. 

So both versions of the installers can released and used IMO.

SylvainCorlay:
> as i mentionned, i've personally had trouble with the mamba1 -> mamba2 transition on a few of my machines quite literally yesterday. It has just been a very tiring transition that has made me consider dropping mamba from Miniforge3 alltogether (I don't want to do it, but I think it is something to consider given the amount of recent churn it has caused me personally). https://github.com/conda-forge/miniforge/issues/730#issue-2886133892

The transition from Mamba 1 to Mamba 2 has indeed been tiring, even for the maintenance team. However, the Mamba 2 codebase is significantly simpler and readable, more thoroughly tested, and more robust than its predecessor. It should be much simpler to maintain in the long run.

The ABI breakage of `fmt` that occured between our last RC and the final release for this patch version is sheer bad luck - but the team is on the deck working on it, and improving the constructor.

Besides, there will be second order benefits to using Mamba 2 over Mamba 1, such as support for OCI registries, mirrors, etc.

hmaarrfk:
if we are rebuilding for 

* mamba2 + fmt 11.1 workaround


i would rather first build for

* mamba1 + fmt 11.1 workaround
* mamba2 + fmt 11.1 workaround

in quick succession.

Is it ok if i merge https://github.com/conda-forge/miniforge/pull/741 then bring back the mamba2 changes?

SylvainCorlay:
> i would rather first build for
> 
> * mamba1 + fmt 11.1 workaround
> * mamba2 + fmt 11.1 workaround
> 
> in quick succession.
> 
> Is it ok if i merge #741 then bring back the mamba2 changes?

I don't have an objection.

hmaarrfk:
Just built a pre-release tag with 11.0

PR #776, opened 2025-04-28
Update to mamba 2.1.0

jjerphan:
I wonder whether this could be automated.

I am only updating mamba because conda 25.4.0 is not out yet.

cc @hmaarrfk.


jjerphan:
I agree.

Closing in preference of https://github.com/conda-forge/miniforge/pull/777 since the scope changed.

hmaarrfk:
for others, we try to add testing to the build process, but it doesn't really mimic a "human" using the tool, rather it mimics our "expert understanding of how the tool should work in a CI".

PR #792, opened 2025-05-15
Update user path to condabin

jslatane:
Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ the Miniforge Prompt.
 conda init
 ```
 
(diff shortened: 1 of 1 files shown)


hmaarrfk:
no, that was your chosen installation location. the other location is still valid.

jslatane:
Ah, gotcha. Sorry, I assumed this was where it would always try to put a user-installation since this is what popped up by default with the installer when you leave the option for "Install for:" to "Just Me (recommended)". Is there something that controls whether the user installation goes into AppData/Local vs right into the user folder?

PR #879, opened 2026-05-14
Add clarification regarding shell initialization for the macOS pkg installer

psobolewskiPhD:
Checklist

I was helping an end user, for the first time using the PKG installer on macOS instead of the script installer. It went great, but their shell wasn't initialized. I was confused, because I didn't see any prompt for it and eventually realized it was in the Customize option.
So here I just update the README to make that more clear.
Note: it might be worth making that the default in the GUI installer?

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,8 @@ or [Intel](https://github.com/conda-forge/miniforge/releases/latest/download/Min
 and follow the steps on screen by pressing Continue.
 
 By the 4th screen, you can choose a different installation path by clicking on "Change Install Location".
-Other options may be available behind the "Customise" button.
+Other options may be available behind the "Customise" button. Importantly, you can choose to have the intaller
+perform conda intialization for your shell.
 
 Once ready, click on Install. If everything went according to plan, the Summary page will report success.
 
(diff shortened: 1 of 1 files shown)


psobolewskiPhD:
This is also related to: https://github.com/conda-forge/conda-forge.github.io/issues/2834
getting the PKG installer to be on the conda-forge.org/download page instead of the script.

README.md
```diff
@@ -184,7 +184,8 @@ or [Intel](https://github.com/conda-forge/miniforge/releases/latest/download/Min
 and follow the steps on screen by pressing Continue.
 
 By the 4th screen, you can choose a different installation path by clicking on "Change Install Location".
-Other options may be available behind the "Customise" button.
+Other options may be available behind the "Customise" button. Importantly, you can choose to have the intaller
+perform conda intialization for your shell.
```
psobolewskiPhD:
```suggestion
Other options may be available behind the "Customise" button. Importantly, you can choose to have the installer
perform conda initialization for your shell.
```

jaimergp:
> Note: it might be worth making that the default in the GUI installer?

Let's open a issue to discuss that, it led to some users being surprised in the past. We can also customise the text in the installer itself in upcoming PRs if necessary.

jaimergp (approved):

PR #907, opened 2026-09-19
[docs] Fix release table for multiple extensions, some accessibility tweaks

bollwyvl:
As always, thanks so much for maintaining this!

Changes:
- [x] fix `rowspan` to account for multiple file extensions per architecture
  - > for `.pkg` and `.dmg` (and, in the future, `.msi` and `.exe`)
- [x] make link hit targets bigger per [WAI SC 2.5.8](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html)
- [x] move GitHub links out of `summary` which made expanding a release's table harder
- [x] clean up with some linters, type hints

Screens:

- before and after
 - image


bollwyvl:
Of note, the latest ~88 assets already contain a `"digest": "sha256:.."`. So we could skip at least that many fetches, and am happy to do so. Perhaps these could be backfilled somehow for the historic ones to save a whole bunch?

Alternately, an authenticated graphql query could probably get us just what is needed in many fewer queries.

bollwyvl:
... just checked, and while we can get a much smaller payload, old releases still don't have `digest` in GraphQL:

```python
{
    'isDraft': False,
    'isPrerelease': True,
    'name': 'Miniforge3-4.7.12-0',
    'publishedAt': '2019-11-21T04:46:06Z',
    'releaseAssets': {'nodes': [{'digest': None,
                                'downloadUrl': 'https://github.com/conda-forge/miniforge/releases/download/4.7.12-0/Miniforge3-4.7.12-0-Linux-aarch64.sh',
                                'name': 'Miniforge3-4.7.12-0-Linux-aarch64.sh',
                                'size': 52313522},
                                {'digest': None,
                                'downloadUrl': 'https://github.com/conda-forge/miniforge/releases/download/4.7.12-0/Miniforge3-Linux-aarch64.sh',
                                'name': 'Miniforge3-Linux-aarch64.sh',
                                'size': 52313522},
                                {'digest': None,
                                'downloadUrl': 'https://github.com/conda-forge/miniforge/releases/download/4.7.12-0/Miniforge3-4.7.12-0-Linux-aarch64.sh.sha256',
                                'name': 'Miniforge3-4.7.12-0-Linux-aarch64.sh.sha256',
                                'size': 105}]}
},

```

hmaarrfk:
> Of note, the latest ~88 assets already contain a "digest": "sha256:..". So we could skip at least that many fetches, and am happy to do so. Perhaps these could be backfilled somehow for the historic ones to save a whole bunch?



does this mean that github now provides the sha256 and we shouldn't include it as a separate file?

hmaarrfk:
ifyou think this is good to merge, let me know so i can click merge

bollwyvl:
> good to merge

I think it's good enough, given the current page looks broken :sweat_smile:  

> we shouldn't include it as a separate file?

I'd keep them for now, provided there's no convenient way to get them into a single, more canonical `SHA256SUMS` file.

